PR #3, opened 2019-04-15
Feature/spk 26 reader writer main menu

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/activity/MainActivity.kt
```diff
@@ -1,37 +1,89 @@
...
         }
 
-        deleteButton.setOnClickListener {
+        viewModel.deleteFormEvent().observeEvent(this) {
             val intent = Intent(this, DeleteFormActivity::class.java)
             startActivity(intent)
         }
 
+        viewModel.readKeyEvent().observeEvent(this) {
+            Toast.makeText(this, "TODO", Toast.LENGTH_SHORT).show()
```
cquemin: Ticket number here as well. We don't want to still have todos in the code at the end.

app/src/main/java/com/simplifire/yotinfcclient/activity/MainActivity.kt
```diff
@@ -1,37 +1,89 @@
...
+
+        viewModel.wipeKeyEvent().observeEvent(this) {
+            Toast.makeText(this, "TODO", Toast.LENGTH_SHORT).show()
+        }
+    }
+
+    private fun showWriteKeyErrorDialog() {
+        AlertDialog.Builder(this)
+            .setTitle(R.string.you_must_create_form_msg)
+            .setNeutralButton(android.R.string.ok, null).show()
```
cquemin: formatting.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/MainVm.kt
```diff
@@ -0,0 +1,58 @@
...
+) : ViewModel() {
+
+    private val writeKeyErrorEvent = MutableLiveData<Event<Boolean>>()
+    private val writeKeyEvent = MutableLiveData<Event<Boolean>>()
+    private val readKeyEvent = MutableLiveData<Event<Boolean>>()
+    private val wipeKeyEvent = MutableLiveData<Event<Boolean>>()
+    private val createFormEvent = MutableLiveData<Event<Boolean>>()
+    private val deleteFormEvent = MutableLiveData<Event<Boolean>>()
+
+    fun writeKeyErrorEvent() = writeKeyErrorEvent
```
cquemin: Why not have a `public val writeKeyErrorEvent` as opposed to a private val and a getter function?

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/MainVm.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    fun wipeKey() {
+        wipeKeyEvent.value = Event(true)
+    }
+
+    fun createForm() {
+        createFormEvent.value = Event(true)
+    }
+
+    fun deleteForm() {
```
cquemin: I assume this will change most likely once you integrate with the lib?

AlexMisiulia: @cquemin I don't understand exactly what do you mean

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/MainVm.kt
AlexMisiulia: @cquemin Just don't want want to expose `Mutable` object from ViewModel. Changed the code

app/src/main/java/com/simplifire/yotinfcclient/activity/MainActivity.kt
AlexMisiulia: @cquemin Done.

app/src/main/java/com/simplifire/yotinfcclient/activity/MainActivity.kt
AlexMisiulia: @cquemin Done.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/MainVm.kt
AlexMisiulia: @cquemin Done.

cquemin: What I meant is I believe that
```
private val writeKeyErrorEvent = MutableLiveData<Event<Boolean>>()
fun writeKeyErrorEvent() = writeKeyErrorEvent
```
is equivalent to 
`val writeKeyErrorEvent = MutableLiveData<Event<Boolean>>()`
it is a val so you can't modify it and you can still access it because it is public.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/MainVm.kt
cquemin: I meant those method only send event at the moment. Eventually they will be wired to real call. Or you want to keep this event based implementation ?

AlexMisiulia: @cquemin these events are just for button actions on the main screen. I think they will stay with this implementation.

cquemin: Ah ok cool. Let's see how it fits in the rest when merged then.

PR #4, opened 2019-04-16
Add field validation, beautified UI at CreateFormBase screen

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Date
+import javax.inject.Inject
+
+class CreateFormBaseVm @Inject constructor(
+    private val formDao: FormDao,
+    private val androidExecutor: AndroidExecutor,
+    private val dateUtils: DateUtils
+) : ViewModel() {
+
+    private val showDatePickerLiveData = MutableLiveData<Event<Date>>()
```
cquemin: Same comment as on the first PR this could be `val showDatePickerLiveData = MutableLiveData<Event<Date>>()` as opposed to the combination of` private val` and a `fun`

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
```diff
@@ -0,0 +1,75 @@
...
+    private val continueLiveData = MutableLiveData<Event<BaseFormInfo>>()
+    private val formNameValidationErrorLiveData = MutableLiveData<Event<Int?>>()
+
+    fun showDatePickerEvent() = showDatePickerLiveData
+    fun keyExpiryDate() = keyExpiryTextLiveData
+    fun continueEvent() = continueLiveData
+    fun formNameValidationErrorEvent() = formNameValidationErrorLiveData
+
+    fun displayDate(year: Int, month: Int, day: Int) {
+        val beautifiedDate = dateUtils.format(year, month, day, DD_MM_YY_FORMAT)
```
cquemin: I would name this constant `DATE_FORMAT_DD_MM_YY` so that if I don't know what format is available I can figure it out by typing `DATE_FORMAT`

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    fun onContinueClicked(projectName: String, formName: String, keyExpiryDate: String) {
+        if(formName.isEmpty()) {
+            formNameValidationErrorLiveData.value = Event(R.string.form_name_must_be_not_empty)
+            return
+        }
+
+        val resultProjectName = if(projectName.isEmpty()) DEFAULT_PROJECT_NAME else projectName
+
+        androidExecutor.runAsync {
```
cquemin: Definitely should use coroutine (as mentioned in the first PR). As this is similar but not as efficient.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
```diff
@@ -0,0 +1,75 @@
...
+                    formNameValidationErrorLiveData.value = Event(R.string.duplicated_form)
+                } else {
+                    continueLiveData.value = Event(BaseFormInfo(resultProjectName, formName, keyExpiryDate))
+                }
+            }
+        }
+    }
+
+    fun onProjectNameChanged() {
+        if(formNameValidationErrorLiveData.value != null && formNameValidationErrorLiveData.value?.peekContent() != null) {
```
cquemin: I believe you can remove the first condition and your if statement becomes: 
` if(formNameValidationErrorLiveData.value?.peekContent() != null) {`

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
```diff
@@ -0,0 +1,75 @@
...
+            return
+        }
+
+        val resultProjectName = if(projectName.isEmpty()) DEFAULT_PROJECT_NAME else projectName
+
+        androidExecutor.runAsync {
+            val existingForm = formDao.getForm(formName, resultProjectName)
+
+            androidExecutor.runOnUiThread {
+                if (existingForm != null) {
```
cquemin: This condition seems to be always true, why are you testing it?

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormBaseFragment.kt
```diff
@@ -0,0 +1,133 @@
...
+        super.onViewCreated(view, savedInstanceState)
+
+        Injector.appComponent.inject(this)
+        initView()
+        initViewModel()
+    }
+
+    private fun initViewModel() {
+        viewModel = ViewModelProviders.of(this, viewModelFactory).get(CreateFormBaseVm::class.java)
+        viewModel.keyExpiryDate().observe(this, Observer {
```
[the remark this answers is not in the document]
AlexMisiulia: Nope, I need to specify java Observer interface, that's why I must write it :(

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormBaseFragment.kt
```diff
@@ -0,0 +1,133 @@
...
+            showDatePicker(it)
+        }
+
+        viewModel.continueEvent().observeEvent(this) {
+            val detailsFragment = DetailsFormTemplateFragment.newInstance(it.formName, it.projectName, it.keyExpiryDate)
+            listener?.navigateTo(detailsFragment)
+        }
+
+        viewModel.formNameValidationErrorEvent().observeEvent(this) {
+            if(it != null) {
```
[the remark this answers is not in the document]
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
AlexMisiulia: I want to expose LiveData to view, instead of Mutable type. Fixed the code

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
AlexMisiulia: I agree with you, but I have this code everywhere. I will fix it after merge all to dev branch (your comment is still there so I won't forget)

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
AlexMisiulia: My misspell it can be null when you didn't create the form with the same name before. Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormBaseVm.kt
```diff
@@ -22,21 +23,21 @@ class CreateFormBaseVm @Inject constructor(
     private val continueLiveData = MutableLiveData<Event<BaseFormInfo>>()
     private val formNameValidationErrorLiveData = MutableLiveData<Event<Int?>>()
 
-    fun showDatePickerEvent() = showDatePickerLiveData
-    fun keyExpiryDate() = keyExpiryTextLiveData
-    fun continueEvent() = continueLiveData
-    fun formNameValidationErrorEvent() = formNameValidationErrorLiveData
+    fun showDatePickerEvent(): LiveData<Event<Date>> = showDatePickerLiveData
```
cquemin: Ok I understand in this case you should use kotlin convention example from [their website](https://kotlinlang.org/docs/reference/coding-conventions.html):
```
class C {
    private val _elementList = mutableListOf<Element>()

    val elementList: List<Element>
         get() = _elementList
}
```
This way this more Kotlin friendly with the rules of naming a backing field with _ and remove the need for 'fake' function since this is exactly why properties exist in Kotlin.

AlexMisiulia: Sounds good, but it creates even more boilerplate code. Anyway, I completed it )

PR #5, opened 2019-04-19
Feature/spk 27 create form details

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
```diff
@@ -25,57 +26,80 @@ private val callback = object: DiffUtil.ItemCallback<FormCustomField>() {
...
+    }
 
-            removeFieldButton.setOnClickListener {
-                val customField = getItem(adapterPosition)
-                onFormFieldActionListener.onRemoveFieldClick(customField)
-            }
+    override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int) {
+        val formField = getItem(position)
+        if(isRawField(formField)) {
+            (holder as RawFieldViewHolder).bind(rawFieldName, rawFieldCharLimit)
```
cquemin: I am not sure I understand why you need to pass these value in the `bind` method consideration your `RawFieldViewHolder` are built with the textWatcher in parameters. You could access the value from the watcher itself. More importantly I don't think the text watchers and the raw values needs to be in the adapter at all.

I think you could create those 2 text watchers and retrieve the internal value from the `RawFieldViewHolder` itself. This would make this class much cleaner and would remove those indirection/update between the adapter and one instance of the view holder.

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
```diff
@@ -77,24 +57,42 @@ class DetailsFormTemplateFragment : Fragment() {
...
                 formNameTextView.text = it.name
-                projectNameTextView.text = it.projectName
-                keyExpireDateTextView.text = it.keyExpireDate
+                projectNameInputTextView.text = it.projectName
+                keyExpireDateInputTextView.text = it.keyExpireDate
 
                 adapter.submitList(it.customFields)
+                scrollScreenToBottom()
+
+                val isRawMode = it.customFields.isNotEmpty() && it.customFields.any { field -> field == null }
```
cquemin: This should be internal to the UiForm object and you should have a method like `hasRawField():Boolean` on the `UiForm` object

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
```diff
@@ -77,24 +57,42 @@ class DetailsFormTemplateFragment : Fragment() {
...
+                rawFieldModeGroup.setVisible(isRawMode)
+                regularModeGroup.setVisible(!isRawMode)
+            }
+        })
+
+        viewModel.isRawMode().observe(this, Observer { isRawMode ->
+            if(isRawMode != null) {
```
cquemin: Do you need the test here ? I believe the type is `Boolean!`

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
```diff
@@ -77,24 +57,42 @@ class DetailsFormTemplateFragment : Fragment() {
...
+                rawFieldModeGroup.setVisible(isRawMode)
+                regularModeGroup.setVisible(!isRawMode)
+            }
+        })
+
+        viewModel.isProjectNameVisible().observe(this, Observer { isProjectNameVisible ->
+            if(isProjectNameVisible != null) {
```
cquemin: Do you need the test here ? I believe the type is `Boolean!`

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
```diff
@@ -77,24 +57,42 @@ class DetailsFormTemplateFragment : Fragment() {
...
+                projectNameInputTextView.setVisible(isProjectNameVisible)
+                projectNameTextView.setVisible(isProjectNameVisible)
```
cquemin: You have a hardcoded success message line 100, please use string reference.

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
```diff
@@ -123,6 +121,30 @@ class DetailsFormTemplateFragment : Fragment() {
         saveFormButton.setOnClickListener {
             viewModel.saveForm()
         }
+
+        saveFieldButton.setOnClickListener {
+            viewModel.saveField(adapter.getRawFieldName(), adapter.getRawFieldCharLimit())
```
cquemin: This type of method should have an object in parameter as opposed to raw type. I feel like a method 
`saveField(field:FormCustomField)` makes perfect sense. And then internally you might access the raw value and that's fine

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
```diff
@@ -0,0 +1,27 @@
+package com.simplifire.yotinfcclient.utils
+
+import android.text.Editable
+import android.text.TextWatcher
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+
+fun View.setVisible(isVisible: Boolean) {
```
cquemin: nice one!

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
```diff
@@ -8,45 +8,39 @@ import com.simplifire.yotinfcclient.entity.Form
...
     private lateinit var currentUiFormState: UiForm
 
-    fun uiForm() : LiveData<UiForm> {
-        return uiFormLiveData
-    }
-
-    fun formSavedEvent() : LiveData<Event<Boolean>> {
-        return formSavedLiveData
-    }
+    fun uiForm() : LiveData<UiForm> = uiFormLiveData
```
cquemin: same comment on split between private val + fun where a public val would do the same

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
```diff
@@ -8,45 +8,39 @@ import com.simplifire.yotinfcclient.entity.Form
...
+    fun formSavedEvent() : LiveData<Event<Boolean>> = formSavedLiveData
+    fun isRawMode() : LiveData<Boolean> = isRawModeLiveData
+    fun isProjectNameVisible(): LiveData<Boolean> = isProjectNameVisibleLiveData
 
     fun saveField(fieldName: String, charLimitString: String) {
         val charLimit = if(charLimitString.isEmpty()) null else charLimitString.toInt()
         val customField = FormCustomField(fieldName, charLimit)
 
-
```
cquemin: Off the back of my comment about the signature of this method, your `CustomFormFieldAdapter` already contains a list of `FormCustomField` so you would not need to recreate the field again here.
Technically this statement `val charLimit = if(charLimitString.isEmpty()) null else charLimitString.toInt()` should be internal to how you construct your `FormCustomField`. Otherwise you rely on this logic to be applied everytime a `FormCustomField` is built. (even if it is only here for now)

app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
AlexMisiulia: It needs for case of elements reusing (very rare but it exists). For example when you scroll down and then return to the raw field you need to restore its content. This is not a problem while we have only one raw field, but for flexibility, I decided to use it.

I like your idea, but I need to expose these values to Fragment. In this case, I need to get the holder by position inside the adapter or save the reference to it anywhere (or even do it from RecyclerView with hard cast, but I think that current solution is less ugly)

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Done, nice suggestion!

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Agree, don't like to call Java classes from java, that's why the type is nullable

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Agree, don't like to call Java classes from java, that's why the type is nullable

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Had the same idea. Initially, I created this object inside the adapter, but then moved it to ViewModel because of some kind of business logic `val charLimit = if(charLimitString.isEmpty()) null else charLimitString.toInt()`.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
AlexMisiulia: Idea is to expose Immutable type just to prevent client code from doing some crazy things with data

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
AlexMisiulia: Agree with you, have the same idea but just didn't move the code to separate class, because it is used only once. In my opinion, this part of the ViewModel responsibility and I am not sure that this logic is enough for creating a new class (because I need to mock it in unit tests). What do you think?

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Agree, after that ui changed and I refactored this in other PR (even the string changed)

app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
cquemin: Like I mentioned later, you should not need those raw value at this point. When a element is selected you can retrieve the value the element itself or tis associated view holder. Also you are sharing textWatcher where they should be shared really they are tight to a specific field which in turn should update a value in your model. Think of it with data bindings in mind where a field would be bound directly to a UI element. If you don't know what I mean I am happy to try to modify it myself.

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
cquemin: If we use only kotlin then we don't have java <-> kotlin issue.

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
cquemin: Yes this business logic should be inside your model, the one held by the adapter and then passed in parameter to `saveField`. It is indeed business logic but located in the wrong place.

app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
AlexMisiulia: I got the idea but not sure about the implementation. Can you modify it if it won't take too much time?

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: I got the idea, but still not sure about implementation details. If you can show it as part of [this](https://github.com/lampkicking/social-impact-keys/pull/5#discussion_r278103938) discussion that would be nice.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
cquemin: If you use your `FormCustomField` in your adapter and in your view model, then you don't have to worry about this anymore. and this logic will just live inside you `FormCustomField`. When it comes to unit test you usually don't mock model object. You create them with the value you need.

app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
```diff
@@ -25,57 +26,80 @@ private val callback = object: DiffUtil.ItemCallback<FormCustomField>() {
...
+private const val RAW_FIELD_TYPE = R.layout.form_custom_raw_field_item_view
+private const val REGULAR_FIELD_TYPE = R.layout.form_custom_field_item_view
 
-    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): CustomFormFieldViewHolder {
-        val view = LayoutInflater.from(parent.context).inflate(R.layout.form_custom_field_item_view, parent, false)
-        return CustomFormFieldViewHolder(view)
-    }
+class CustomFormFieldAdapter :
+    ListAdapter<FormCustomField, RecyclerView.ViewHolder>(callback) {
+    private var rawFieldName = ""
```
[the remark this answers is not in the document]
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
AlexMisiulia: Agree, but I don't think that object creation and field validation is part of FormCustomField. I think that we just can't understand each other. Where do you want to place this logic?
` val charLimit = if(charLimitString.isEmpty()) null else charLimitString.toInt()
   val customField = FormCustomField(fieldName, charLimit)`

This can be some util function or do you want it to be as part of `FormCustomField` class

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Strange, my fault. Replace all useless null checks

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Strange, my fault. Replace all useless null checks

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Strange, my fault. Replace all useless null checks

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
cquemin: did you have a look [at this branch](https://github.com/lampkicking/social-impact-keys/tree/feature/spk-27_create-form-detaails-with-raw) which contains an idea of what I meant:

AlexMisiulia: @cquemin not yet. Today I completed the Read flow. My plan for tomorrow is to complete tasks from code review and I will check your changes

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
cquemin: look at the branch I pushed, there is the example there too.

AlexMisiulia: Thanks, currently working on NFC problems. Will look some time later

app/src/main/java/com/simplifire/yotinfcclient/adapter/CustomFormFieldAdapter.kt
AlexMisiulia: Done, thank you for your idea. Finally, I understand what you mean. I am still don't like that formField is mutable inside the adapter but I can't find any better solution.

app/src/main/java/com/simplifire/yotinfcclient/fragment/CreateFormDetailsFragment.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
AlexMisiulia: Done

PR #7, opened 2019-04-20
Feature/spk 48

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/fragment/FormListFragment.kt
```diff
@@ -93,5 +95,25 @@ class FormListFragment : Fragment() {
         viewModel.error().observeEvent(this) {
             Toast.makeText(activity, R.string.select_form_to_delete_error_msg, Toast.LENGTH_SHORT).show()
         }
+
+        viewModel.descriptionText().observe(this, Observer {
+            if (it != null) {
```
cquemin: No need for null check as the type is `!`

app/src/main/res/values/strings.xml
```diff
@@ -67,4 +67,20 @@ number of days or months, when do you want them to expire?</string>
 
     <!--Create form screen end-->
 
+
+    <!--Delete form screen start-->
+
+    <string name="delete_form_list_item_hint">Please tap the name of the form you want to delete.
+        It is often better to leave old forms in the system for future reference</string>
```
cquemin: should this contain a `\n` or not ?

app/src/main/java/com/simplifire/yotinfcclient/fragment/FormListFragment.kt
AlexMisiulia: Agree, don't like to call Java classes from Kotlin, that's why the type is nullable

app/src/main/res/values/strings.xml
AlexMisiulia: No. This is just to avoid long lines

PR #8, opened 2019-04-20
Feature/spk 49

AlexMisiulia:


app/build.gradle
```diff
@@ -48,6 +45,15 @@ dependencies {
     implementation 'com.facebook.stetho:stetho:1.5.1'
 
     implementation 'com.google.code.gson:gson:2.8.5'
+
+    // test dependencies
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'org.mockito:mockito-core:2.19.0'
```
cquemin: externalise properties in gradle.properties file

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/FormListVm.kt
```diff
@@ -65,14 +65,35 @@ class FormListVm @Inject constructor(
...
+        val selectedFormIndex = currItems.indexOf(item)
+        updatedItems[selectedFormIndex] = item.copy(isSelected = true)
+
+        formListLiveData.value = updatedItems
+    }
+
+    private fun resetSelectedItem(items: List<UiBaseFormItem>) : MutableList<UiBaseFormItem> {
+        val updatedList = ArrayList(items)
+
+        val prevSelectedItem = items.find { it is UiFormItem && it.isSelected } as UiFormItem?
```
cquemin: how about using `items.forEachIndexed` so that you dont have to use `find` and `indexOf` ? not sure it will be better ?

AlexMisiulia: Tried a similar solution but that would find only one element and access it by index. The amount of code is the same in both cases so I leave this one

app/build.gradle
AlexMisiulia: Can we leave for dev branch too (https://github.com/lampkicking/social-impact-keys/pull/1#discussion_r276541336)? I will be happy if you show me a good example of using this technique.

PR #9, opened 2019-04-24
Feature/spk 50

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/utils/CameraHelper.kt
```diff
@@ -7,14 +7,19 @@ import android.graphics.Bitmap
...
 import java.io.IOException
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
+import javax.inject.Inject
+import javax.inject.Singleton
 
-object CameraHelper {
-    val CAMERA_REQUEST_CODE = 1888
+const val CAMERA_REQUEST_CODE = 1888
```
cquemin: This should be split into several parts.
Dealing with an intent to launch the camera should be done in the activity that needs it. We have a camera module that takes picture and uses face detection and everything. For the demo it is not needed but might be worth adding.
Finally the part where the file is saved and transformed should be in an `PhotoInteractor`

app/src/main/java/com/simplifire/yotinfcclient/fragment/TakePhotoFragment.kt
```diff
@@ -59,30 +66,52 @@ class TakePhotoFragment : Fragment() {
     }
 
     private fun initView() {
-        takePhotoActionButton.setOnClickListener {
-            val context = activity ?: return@setOnClickListener
+        continueButton.setOnClickListener {
+            val photoPath = cameraHelper.getPhotoPath()
+            if(photoPath != null)
+            {
```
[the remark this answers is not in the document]
AlexMisiulia: Fixed

app/src/main/java/com/simplifire/yotinfcclient/utils/CameraHelper.kt
AlexMisiulia: Moved the constant to the Fragment. Agree that I can divide this class into parts, but I am not sure that I should do it now

cquemin: I see your point but if you prefer to merge this I won't fight it, but this is the type of things that stayed if not done now.

PR #11, opened 2019-04-24
[ADD] PinCodeFragment with components

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/activity/WriteKeyActivity.kt
```diff
@@ -49,7 +51,17 @@ class WriteKeyActivity :
     }
 
     override fun onFormContinueClick() {
-        // TODO LM_LM
+        val pinCodeHint = getString(R.string.write_key_pin_hint)
+        navigateTo(PinCodeFragment.newInstance(pinCodeHint))
+    }
+
+    override fun onPinCodeEntered(enteredPinCode: String, isSubmitMode: Boolean) {
```
cquemin: Should this be 2 different methods. It would be cleaner to split them in 2 separate callbacks as opposed to have this boolean where we can clearly 2 different flows

app/src/main/java/com/simplifire/yotinfcclient/fragment/PinCodeFragment.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    override fun onCreateView(
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        // Inflate the layout for this fragment
+        return inflater.inflate(R.layout.fragment_pin_code, container, false)
+    }
+
+    override fun onAttach(context: Context) {
```
cquemin: Do we know what device Ken will use for demo. This method was added in Android 24 so we need to make sure his device his at least 24 or above.

app/src/main/java/com/simplifire/yotinfcclient/fragment/PinCodeFragment.kt
```diff
@@ -0,0 +1,127 @@
...
+        super.onAttach(context)
+        if (context is OnPinCodeEnteredListener) {
+            listener = context
+        } else {
+            throw RuntimeException(context.toString() + " must implement ${OnPinCodeEnteredListener::class.java.simpleName}")
```
cquemin: This should be an `IllegalArgumentException`

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+class PinCodeViewModel @Inject constructor() : ViewModel() {
+
+    private var correctPinCode: String? = null
+
+    private val pinCodeCharErrorsLiveData = MutableLiveData<Event<List<Int?>>>()
+    private val pinCodeErrorLiveData = MutableLiveData<Event<Int>>()
+    private val continueEventLiveData = MutableLiveData<Event<PinCodeInfo>>()
+    private val pinCodeHintLiveData = MutableLiveData<String>()
+
+    fun pinCodeCharErrors() : LiveData<Event<List<Int?>>> = pinCodeCharErrorsLiveData
```
cquemin: Same comment on formatting this type of fields

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+                pinCodeErrorLiveData.value = Event(R.string.incorrect_pin_code)
+            }
+
+        } else {
+            val errorList = pinCodeChars.map { if (it.isBlank()) R.string.enter_pin_symbol else null }
+            pinCodeCharErrorsLiveData.value = Event(errorList)
+        }
+    }
+
+    data class PinCodeInfo(val enteredPinCode: String, val isSubmitMode: Boolean)
```
cquemin: This should probably be a sealed class with 2 sub state as opposed to a boolean.

app/src/main/java/com/simplifire/yotinfcclient/fragment/PinCodeFragment.kt
AlexMisiulia: This is Fragment from androidx package, so we won't have any problems with that. I am testing on Nexus 23 API and it works fine

app/src/main/java/com/simplifire/yotinfcclient/activity/WriteKeyActivity.kt
AlexMisiulia: Agree with you, I implemented it the same first time. But now I am thinking about ReadFlow (it will have the same interface but with one more callback when pin code entered incorrectly). I will change it during dev of Read flow. What do you think?

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
AlexMisiulia: Agree. But now I am thinking about ReadFlow (it will have the same interface but with one more callback when pin code entered incorrectly). I will change it during dev of Read flow. What do you think?

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/fragment/PinCodeFragment.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/activity/WriteKeyActivity.kt
cquemin: makes sense put a ticket number then

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
cquemin: As per above

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
cquemin: done where?

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/PinCodeViewModel.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/activity/WriteKeyActivity.kt
AlexMisiulia: Done

PR #12, opened 2019-04-24
Feature/spk 52

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/activity/WriteKeyActivity.kt
```diff
@@ -57,14 +55,20 @@ class WriteKeyActivity :
...
         }
     }
 
+    override fun onNfcActionCompleted() {
+        val title = getString(R.string.write_nfc_key_success_title)
+        val hint = getString(R.string.write_nfc_key_success_hint)
+        navigateTo(SuccessFragment.newInstance(title, hint, true))
+    }
+
     private fun navigateTo(fragment: Fragment, addToBackStack: Boolean = true) {
```
cquemin: This should be put in an abstractActivity in protected

app/src/main/java/com/simplifire/yotinfcclient/fragment/WriteNfcFragment.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.content.Intent
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProviders
+import com.simplifire.yotinfcclient.entity.NfcPersonalFormData
+import com.simplifire.yotinfcclient.nfc.BaseNfcFragment
+import com.simplifire.yotinfcclient.viewmodel.WriteNfcViewModel
+
+class WriteNfcFragment : BaseNfcFragment<WriteNfcViewModel>() {
+
+    private var nfcPersonFormToWrite: NfcPersonalFormData? = null
```
cquemin: Why is this here and not in the view model?

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcFragment.kt
```diff
@@ -0,0 +1,126 @@
...
+    protected lateinit var viewModel: T
+    protected val nfcHelper: NfcHelper = Injector.appComponent.getNfcHelper()
+    protected val viewModelFactory: ViewModelFactory = Injector.appComponent.getViewModelFactory()
+
+    interface OnNfcActionCompletedListener {
+        fun onNfcActionCompleted()
+    }
+
+    abstract fun createViewModel() : T
+    abstract fun init()
```
cquemin: This should read `initViewModel` or something similar as sub classes should not need to infer what this method is initialising.
And abstract methods should be commented: when are they called how to use them etc

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.simplifire.yotinfcclient.entity.WarningInfo
+import com.simplifire.yotinfcclient.entity.WarningState
+import com.simplifire.yotinfcclient.viewmodel.Event
+
+abstract class BaseNfcViewModel : ViewModel() {
+
+    protected lateinit var pinCode: String
+
+    private val nfcHintLiveData = MutableLiveData<String>()
+    protected val finishFragmentLiveData = MutableLiveData<Event<Boolean>>()
```
cquemin: The view model should not have any notion of what type of view it deals with. That is the power of MVVM. It should hold data model I don't think having this type of Event with a boolean is very MVVM friendly

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcDataConverter.kt
```diff
@@ -0,0 +1,34 @@
...
+        // TODO LM_LM replace serialization with protobuf
+        val entityBytes = serialize(entity)
+        return NdefRecord.createMime("text/plain", entityBytes)
+    }
+
+    fun <T> fromNdefRecord(ndefRecord: NdefRecord) : T? {
+        return deserialize<T>(ndefRecord.payload)
+    }
+
+    private fun <T> serialize(entity: T): ByteArray {
```
cquemin: If this is how you use it I really suggest for size purpose to use kotlinx.serialisation as per my first PR comment. I am not sure you will get a very optimized size this way.

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
```diff
@@ -4,123 +4,74 @@ import android.app.Activity
...
-
-        try {
-            // Get the Text
-            text = String(payload, languageCodeLength + 1, payload.size - languageCodeLength - 1, textEncoding)
-        } catch (e: UnsupportedEncodingException) {
-            Log.e("UnsupportedEncoding", e.toString())
+    fun <T> write(intent: Intent?, entity: T, onResult: (Boolean) -> Unit) {
+        val tag: Tag = intent?.getParcelableExtra(NfcAdapter.EXTRA_TAG) ?: return
+
+        threadDispatcher.runAsync {
```
cquemin: That is why I put the comment on the first PR now this is everywhere and I really think it should be changed to use coroutines.

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
```diff
@@ -4,123 +4,74 @@ import android.app.Activity
...
-class NfcHelper {
+private const val TAG = "NfcHelper"
+class NfcHelper @Inject constructor(private val nfcDataConverter: NfcDataConverter,
+                                    private val threadDispatcher: ThreadDispatcher) {
     private lateinit var nfcAdapter: NfcAdapter
     private lateinit var pendingIntent: PendingIntent
     private lateinit var writeTagFilters: Array<IntentFilter>
-    private var myTag: Tag? = null
 
     fun init(activity: Activity) {
```
cquemin: Same comment as for the CameraHelper. Intent management should be split from data manipulation. Data manipulation is a Use case, i.e an Interactor. You should have something like `NFCReadInteractor` and `NFCWriteInteractor`

app/src/main/res/layout/fragment_nfc.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
cquemin: formatting

app/src/main/java/com/simplifire/yotinfcclient/fragment/WriteNfcFragment.kt
AlexMisiulia: I was rushing the last 2 days:( This is because when viewModel receive the information need to schedule it's writing when nfc key will be connected with device. Firstly I implemented this schedule mechanism inside `NfcHelper`, but I didn't like that approach. This one I also don't like but don't have better ideas. Maybe you can suggest something useful?

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcDataConverter.kt
AlexMisiulia: Need to check it

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
AlexMisiulia: Agree, will change. Not much work to replace it everywhere. Need to complete prototype in time and after that will use coroutines

app/src/main/res/layout/fragment_nfc.xml
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
AlexMisiulia: Agree, need time to change

app/src/main/java/com/simplifire/yotinfcclient/fragment/WriteNfcFragment.kt
cquemin: Fair enough. I believe these are data model and therefore should live inside the ViewModel and then the view should subscribe to the correct LiveData to then trigger/update anything it needs, as you did with the other type of model. In this example the NfcHelper should only be a means to end, not where the data lives and not where the condition to trigger its usage either. This should all be driven from the viewModel exposing the right liveData

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
cquemin: I hear you and that is why I mentioned on the first PR to avoid this type of problem where we build on technical debt. I understand you wanted to wait to address them. But we will end up with technical debt on a brand new project. Really not ideal.

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
AlexMisiulia: @cquemin refactored it. I moved all logic to a class `NfcRepository`. We can discuss a bit more naming, but I didn't use clean arch components in this project so I don't think that this is a good idea to start using it here. If you have a practice like that I need to rewatch my whole approach. Just want the code to look consistently

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
AlexMisiulia: Fully agree, not ideal at all, but I had a strict deadline and I wasn't sure that I can complete my work in time. I will refactor it anyway

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcViewModel.kt
AlexMisiulia: Agree about naming. But I can't understand what is the alternative for `<Event<Boolean>> ` type?

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcFragment.kt
AlexMisiulia: @cquemin this method can be used not only for `ViewModel` initialization. It just told view that at this stage it needs to initialize all components which it needs for work

app/src/main/java/com/simplifire/yotinfcclient/fragment/WriteNfcFragment.kt
AlexMisiulia: Changed this code, thanks for the comment!

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcFragment.kt
cquemin: in any case it should at the very least be documented

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcViewModel.kt
cquemin: I am just wondering if there is a way to model this need differently than a boolean. This weak and not very dev friendly either. Maybe sealed class status or something like that would be more self explantory and can also cover more state : INIT, LOADING, LOADED, FINISHED. What do you think?

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
cquemin: no worries I understand having the demo app up and running is the priority

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcViewModel.kt
AlexMisiulia: I understood your idea. But I think these states are for another approach. 

I have thought to try approach when we have one liveData and object that stores entirely screen data like NfcViewState. Here is the link to this approach: https://medium.com/@trionkidnapper/viewmodel-and-kotlin-data-class-7d3a3b854805.

This made code much more readable but creates another problem: every time you change the any variable in view state it will render whole screen. I guess that we won't have perfomance issues with that. I can try this approach on one screen and we will check what looks better. What do you think? I think it will be separate PR

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcDataConverter.kt
AlexMisiulia: UPD: I investigated it and found that kotlinx.serialization have only Json and Protobuf support. Protobuf doesn't support nulls at this moment so I can't use it without some hacks.

This code will be replaced with your library, so maybe we can just skip it for now. What do you think?

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcHelper.kt
AlexMisiulia: This code if refactored. I moved all NFC logic to NfcRepository

app/src/main/java/com/simplifire/yotinfcclient/nfc/BaseNfcFragment.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/activity/WriteKeyActivity.kt
AlexMisiulia: Not sure about that because this method just removes boilerplate code of passing fragmentContainerId argument everywhere. Otherwise, it is just a call of a utility function. I am not sure that it is a candidate for fun in AbstractActivity. What do you think?

PR #13, opened 2019-04-26
SPK-52 [ADD] ReadKey screen with components

AlexMisiulia:
@cquemin I need some time to review the code by myself. Let's follow the rule: while PR is draft it means that it is not ready for your review - I will review it by myself. After I completed that I will change status.


app/src/main/java/com/simplifire/yotinfcclient/viewmodel/ReadNfcViewModel.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    private val _pinCodeIncorrectLiveData = MutableLiveData<Event<Boolean>>()
+    val pinCodeIncorrectLiveData: LiveData<Event<Boolean>>
+        get() = _pinCodeIncorrectLiveData
+
+    private var _readCompletedLiveData = MutableLiveData<Event<Boolean>>()
+    val readCompletedLiveData: LiveData<Event<Boolean>>
+        get() = _readCompletedLiveData
+
+    override fun onSuccessInit() {
```
AlexMisiulia: Need to make this method with a default implementation in the base class

AlexMisiulia: Done

PR #15, opened 2019-04-26
Feature/spk 76

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/fragment/FormListFragment.kt
```diff
@@ -54,10 +55,10 @@ class FormListFragment : Fragment() {
 
     override fun onAttach(context: Context) {
         super.onAttach(context)
-        if (context is OnFormSelectedListener) {
+        if (context is OnFormActionListener) {
             listener = context
         } else {
-            throw RuntimeException("$context must implement OnFormSelectedListener")
+            throw RuntimeException("$context must implement OnFormActionListener")
```
AlexMisiulia: Replace hardcoded interface name with the class name

AlexMisiulia: Done

PR #20, opened 2019-04-29
SPK-86 [REF] store data read from NFC key, write updated info to nfc …

AlexMisiulia:
…key back. Handle case that during exception during writing updated info to key old data erased from it


AlexMisiulia:
Reviewed

PR #22, opened 2019-05-02
Feature/spk 90

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/adapter/FormFieldAdapter.kt
```diff
@@ -31,27 +30,15 @@ class FormFieldAdapter : ListAdapter<UiPersonalFormItem, FormFieldAdapter.FieldV
...
     }
 
     class FieldViewHolder(override val containerView: View)
         : RecyclerView.ViewHolder(containerView), LayoutContainer {
 
-        fun bind(formField: UiPersonalFormItem, isFirst: Boolean, isLast: Boolean) {
+        fun bind(formField: UiPersonalFormItem) {
             fieldNameTextView.text = formField.fieldName
             fieldValueTextView.text = formField.fieldValue
-
```
cquemin: Did you see how many time you had to do the same modification. This is usually a sign that there is too much duplication and the logic should be mutualised if possible

app/src/main/java/com/simplifire/yotinfcclient/fragment/PersonalFormFragment.kt
```diff
@@ -45,6 +46,8 @@ class PersonalFormFragment : Fragment() {
 
     private fun initView() {
         fieldsRecyclerView.layoutManager = LinearLayoutManager(requireActivity(), RecyclerView.VERTICAL, false)
+        val borderDecorator = createBorderDecorator(requireActivity())
```
cquemin: Same comment about code duplication

app/src/main/java/com/simplifire/yotinfcclient/adapter/FormFieldAdapter.kt
AlexMisiulia: I removed duplication to separate class. Or maybe I can't understand you

app/src/main/java/com/simplifire/yotinfcclient/fragment/PersonalFormFragment.kt
AlexMisiulia: Not sure that I understand you. This is just borderDecorator creation. Do you mean this line is code duplication?

app/src/main/java/com/simplifire/yotinfcclient/adapter/FormFieldAdapter.kt
cquemin: No you are right you what's left is not duplicated my bad, I mis read.

app/src/main/java/com/simplifire/yotinfcclient/fragment/PersonalFormFragment.kt
cquemin: I was saying the `initView` method in `PersonalFormFragment` and in `FormDetailFragment` looks very similar.

AlexMisiulia: Yes, but In my opinion, this is the case where copy pasting 3 lines of code
```
        fieldsRecyclerView.layoutManager = LinearLayoutManager(requireActivity(), RecyclerView.VERTICAL, false)
        val borderDecorator = createBorderDecorator(requireActivity())
        fieldsRecyclerView.addItemDecoration(borderDecorator)
```

 is better than create abstraction like `initBorderRecyclerView`

cquemin: If that is the only logic in common then i agree. It was way for me to raise the question: does it make sense to put the logic in common.

PR #25, opened 2019-05-02
Feature/spk 56

AlexMisiulia:


app/src/main/res/values/styles.xml
```diff
@@ -144,6 +144,19 @@
...
 
+    <style name="CheckBox" parent="android:Widget.Material.Light.CompoundButton.CheckBox">
+        <item name="android:layout_width">40dp</item>
+        <item name="android:layout_height">40dp</item>
+        <item name="android:button">@null</item>
+        <item name="android:background">?android:attr/listChoiceIndicatorMultiple</item>
+        <item name="android:theme">@style/CheckBoxTheme</item>
+    </style>
+
+    <style name="CheckBoxTheme" parent="AppTheme">
```
cquemin: It should be a style. You don't really apply theme to a ui widget.

AlexMisiulia: No, in line 152 I should use it as a theme. This is the only way I can change the border color for CheckBox :(

PR #26, opened 2019-05-03
Feature/spk 55

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcRepository.kt
```diff
@@ -1,27 +1,56 @@
 package com.simplifire.yotinfcclient.nfc
 
 import android.nfc.NdefMessage
+import android.nfc.NdefRecord
 import android.nfc.tech.Ndef
 import android.nfc.tech.NdefFormatable
+import android.util.Log
```
AlexMisiulia: Remove useless import

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
```diff
@@ -16,10 +16,12 @@ fun ViewGroup.inflateAdapterView(layoutId: Int): View {
     return LayoutInflater.from(context).inflate(layoutId, this, false)
 }
 
-fun showInfoDialog(messageResId: Int, context: Context) {
+fun showInfoDialog(messageResId: Int, context: Context, neutralButtonCode: ()->Unit = {}) {
```
AlexMisiulia: Fix formatting

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcWriteError.kt
```diff
@@ -1,6 +1,6 @@
 package com.simplifire.yotinfcclient.nfc
 
-class NfcWriteError(val type: NfcWriteErrorType) : Exception()
+class NfcWriteError(val type: NfcWriteErrorType) : RuntimeException()
```
cquemin: If it is an exception, why naming it `xxxError`?

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        setLoading(true)
+        wipeNfcKey(nfcNdefData)
+    }
+
+    private fun wipeNfcKey(nfcNdefData: NfcNdefData) {
+        viewModelScope.launch {
+
+            val nfcPersonalData = nfcRepository.readAsync<NfcPersonalFormData>(nfcNdefData)
```
cquemin: It is no async actually. This is a suspending method so it will suspend the current thread until it completes.
That is the idea behing using coroutine the code becomes more functional and does not need the notion of async to be explicit.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+                nfcRepository.wipeAsync(nfcNdefData)
+            } catch (nfcWriteException: NfcWriteError) {
+                nfcError = nfcWriteException
+            }
+
+            handleNfcWipeResult(nfcError)
+        }
+    }
+
+    private fun handleNfcWipeResult(nfcError: NfcWriteError?) {
```
cquemin: Confusing method name: it reads handle result but deals only with an error.
I'd suggest to remove this method, to call `setNfcError(nfcWriteException)` in the catch above and add `_navigateToSuccessScreen.value = Event(true)` after the `try/catch` bloc.
Having an optional error which means success if not there goes against safe type. Since you are not doing anything with the exception (you are not even logging which you should) I would suggest that `nfcRepository.wipeAsync(nfcNdefData)` should not raise an exception but instead return a sealed class for success or Error. This way you'd have a `when` statement here and a much cleaner management of your states.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+        setLoading(false)
+        if (nfcError == null) {
+            _navigateToSuccessScreen.value = Event(true)
+        } else {
+            setNfcError(nfcError)
+        }
+    }
+
+    private fun setNfcError(nfcWriteException: NfcWriteError) {
```
cquemin: this method is again named counter intuitively. I thought this was setting a state, but this is showing a message to the user. Therefore should be named with a name that reflects this.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+        when (nfcWriteException.type) {
+            NfcWriteErrorType.IO_EXCEPTION -> setWarningInfo(defaultErrorInfo)
+            NfcWriteErrorType.NFC_TAG_NOT_ENOUGH_SIZE -> setDialog(R.string.write_key_not_enough_space_error_msg)
+            NfcWriteErrorType.NFC_TAG_NOT_WRITABLE -> setDialog(R.string.write_key_not_writable_error_msg)
+            NfcWriteErrorType.UNKNOWN_NFC_TECH -> setDialog(R.string.write_key_not_writable_error_msg)
+        }
+    }
+
+    private fun setDialog(stringRes: Int) {
```
cquemin: on comment here. Weird naming.

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcRepository.kt
```diff
@@ -1,27 +1,65 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.withContext
 import java.io.IOException
 import javax.inject.Inject
 import javax.inject.Singleton
 
 private const val TAG = "NfcRepository"
 
+
+interface NfcRepository {
```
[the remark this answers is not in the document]
AlexMisiulia: Agree, done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
AlexMisiulia: This is async under the hood. It uses `withContext(ioDispatcher) { }` threading, so it will be async.

AlexMisiulia: I saw an example like this inside one of the kotlin repo, samples.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
AlexMisiulia: I thought about it, but viewmodel didn't show anything to user, it stores the state that the view observes. From that point of view the naming is correct. I watched samples of open source MVVM apps and someone names methods with 'display/show' prefix, someone prefers 'set' prefix. I think that 'display/show' is more intuitive but is not very clean for ViewModel meaning. What do you think?

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcWriteError.kt
AlexMisiulia: Agreed. Will do it in another branch to avoid conflicts. Link: https://github.com/lampkicking/social-impact-keys/pull/27/files#r282476433

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
AlexMisiulia: I like the first idea.  The second one adds another level of abstraction in my opinion.

AlexMisiulia: Done it in https://github.com/lampkicking/social-impact-keys/pull/27 to avoid merge conflicts

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
cquemin: Yeah I see your point. It makes good sense. I will check with the rest of the android team to see what they think.

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
AlexMisiulia: as per above comment

app/src/main/java/com/simplifire/yotinfcclient/nfc/NfcWriteError.kt
AlexMisiulia: Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/WipeNfcViewModel.kt
cquemin: SetXXX is probably ok. We use Handlexxxxx but it does not matter too much

PR #28, opened 2019-05-09
Feature/spk 96. Add max field name validation

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
```diff
@@ -48,3 +52,30 @@ interface DefaultTextWatcher : TextWatcher {
...
+    override fun filter(
+        source: CharSequence,
+        start: Int,
+        end: Int,
+        dest: Spanned,
+        dstart: Int,
+        dend: Int
+    ): CharSequence? {
+        try {
+            val input = Integer.parseInt(dest.toString() + source.toString())
```
cquemin: Should this be done with a regular expression instead ?

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
```diff
@@ -48,3 +52,30 @@ interface DefaultTextWatcher : TextWatcher {
...
+            if (isInRange(min, max, input)) {
+                return null
```
cquemin: If it is in range you return `null` ? is that correct? If so this method should be a boolean since it only returns 2 type of values. And this is usually what you would expect from a `filter` method.

app/src/main/res/values/integer.xml
```diff
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <integer name="pin_code_input_max_length">1</integer>
+    <integer name="form_field_name_max_length">50</integer>
+    <integer name="form_char_limit_max_number">4000</integer>
+    <integer name="form_name_max_length">50</integer>
```
cquemin: This is not stored on the key why having such a small max length ?

app/src/main/res/values/integer.xml
```diff
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <integer name="pin_code_input_max_length">1</integer>
+    <integer name="form_field_name_max_length">50</integer>
+    <integer name="form_char_limit_max_number">4000</integer>
```
cquemin: What is this used for?

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
```diff
@@ -48,3 +52,30 @@ interface DefaultTextWatcher : TextWatcher {
     }
 }
 
+class InputFilterMinMax(private var min: Int, private var max: Int) : InputFilter {
```
cquemin: This should be named `InputMinMaxValidator` as opposed to `filter` because it looks like your filter method is actually validating the content as opposed to filtering it.

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
AlexMisiulia: Not sure, this is just usual concatenation. Why do we need regexp here?

app/src/main/res/values/integer.xml
AlexMisiulia: Max number for "charLimit" field

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
AlexMisiulia: This is confusing Android Api, but I implement interface InputFilter, that's why it should be filter

app/src/main/res/values/integer.xml
AlexMisiulia: I am discussing it with Simon now.

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
AlexMisiulia: This is Android Api, I am still confused about this method.

Part of the documentation:

Return the CharSequence that you would like to have placed there instead, including an empty string if appropriate, or null to accept the original replacement.

app/src/main/res/values/integer.xml
AlexMisiulia: Simon said that max_length 50 is ok

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
cquemin: not much you are right.

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
cquemin: Oh right.... in that case ok :)

app/src/main/res/values/integer.xml
cquemin: and 4k is not too much ? what does that include in those 4k?

AlexMisiulia: Check comment for this issue please https://lampkicking.atlassian.net/browse/SPK-96

PR #33, opened 2019-05-10
Feature/spk 109. Create form flow: replace field specific dialog errors with in-line messages

AlexMisiulia:


app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
```diff
@@ -20,6 +20,20 @@ fun ViewGroup.inflateAdapterView(layoutId: Int): View {
     return LayoutInflater.from(context).inflate(layoutId, this, false)
 }
 
+fun TextView.setErrorRes(errorRes: Int?) {
```
cquemin: This should use a `@StringRes` annotation

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
```diff
@@ -26,43 +26,40 @@ class CreateFormDetailsVm @Inject constructor(private val formDao: FormDao,
...
-    val fieldErrorLiveData: LiveData<Event<Int>>
-        get() = _fieldErrorLiveData
-
     private lateinit var currentUiFormState: UiForm
 
     fun saveField(rawField: UiFormCustomRawField) {
-        if(!validateField(rawField)) return
+        val validatedField = validateField(rawField)
+
+        val newField = if (!validatedField.hasErrors()) {
```
cquemin: This is slightly counter intuitive. You have a variable named `validatedField` (which comes as result of a method named `validateField `, which makes perfect sense. But then you checked if it has errors ? so it is not a validatedField if it does contain error right? and then the result of this block return a field.

I think validateField should also check internally if there are errors and return the correct Field type: raw or regular.

app/src/main/java/com/simplifire/yotinfcclient/utils/ViewUtils.kt
AlexMisiulia: Nice suggestion, thanks! Done

app/src/main/java/com/simplifire/yotinfcclient/viewmodel/CreateFormDetailsVm.kt
AlexMisiulia: Sounds good. Done

PR #47, opened 2019-05-15
Feature/spk 62.  Create Demo Main Page

AlexMisiulia:


app/src/main/java/com/yoti/mobile/android/social/keys/nfc/features/read/ReadNfcViewModel.kt
```diff
@@ -102,7 +102,8 @@ class ReadNfcViewModel @Inject constructor(
             if (isSuccessful) {
                 // need to store in memory data from nfc key to show previous lastVisitDate
                 personalFormRepository.personalForm = nfcPersonalFormData.personalForm
-                _readCompletedLiveData.value = Event(true)
+                _readCompletedLiveData.value =
```
AlexMisiulia: Remove strange formatting

AlexMisiulia: Done

.idea/gradle.xml
```diff
@@ -9,6 +9,8 @@
           <set>
             <option value="$PROJECT_DIR$" />
             <option value="$PROJECT_DIR$/app" />
+            <option value="$PROJECT_DIR$/common" />
```
cquemin: Can you add `.idea/ `to `.gitignore` file ?

common/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        }
+    }
+
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
```
cquemin: This line can be removed.

common/src/main/java/com/yoti/mobile/android/social/common/di/ViewModelFactory.kt
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Inject
+import javax.inject.Provider
+import javax.inject.Singleton
+
+@Singleton
+class ViewModelFactory @Inject constructor(private val viewModels: MutableMap<Class<out ViewModel>, Provider<ViewModel>>) :
+    ViewModelProvider.Factory {
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T : ViewModel> create(modelClass: Class<T>): T {
```
cquemin: Have you thought of using `reified` here ? I think it might allow you to avoid suppressing the unchecked cast

common/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
```
cquemin: this file can be removed.

demo/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    implementation project(':common')
+
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
```
cquemin: this line can be removed.

demo/src/main/java/com/yoti/mobile/android/social/demo/SocialKeysDemoApplication.kt
```diff
@@ -0,0 +1,17 @@
+package com.yoti.mobile.android.social.demo
+
+import android.annotation.SuppressLint
+import android.app.Application
+import android.content.Context
+
+class SocialKeysDemoApplication : Application() {
+    companion object {
+        @SuppressLint("StaticFieldLeak")
+        lateinit var instance: Context
```
cquemin: Why are you doing this ? You can always retrieve your application from the context and since your viewModel lives in a fragment or in an activity you can always pass down the chain your context.
I don't think this is a good design decision to store statically a reference of the application since it is already guarantee to always have one instance.

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
```diff
@@ -0,0 +1,49 @@
+package com.yoti.mobile.android.social.demo.repository
+
+import com.yoti.mobile.android.social.demo.R
+import com.yoti.mobile.android.social.demo.entity.Organization
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
cquemin: Why does this have to be a singleton?

.idea/gradle.xml
AlexMisiulia: Done

common/build.gradle
AlexMisiulia: Done

common/src/main/java/com/yoti/mobile/android/social/common/di/ViewModelFactory.kt
AlexMisiulia: Can't use it because I need to use the inline function, can't use the inline function because I need to access private val :(

common/src/main/res/values/strings.xml
AlexMisiulia: Done

demo/build.gradle
AlexMisiulia: Done

demo/src/main/java/com/yoti/mobile/android/social/demo/SocialKeysDemoApplication.kt
AlexMisiulia: Just need to have quick access to it. I use it primarily in `Injector` or `Constants` to retrieve the resource or create the `AppComponent`. I don't pass Context to the ViewModel where can avoid for ease unit testing

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
AlexMisiulia: In order to don't recreate this object while passing to different `ViewModel`

cquemin: Yes but why are you worried that this will be recreated ? is it for performance issue ?
This object does not create its own DB nor does it do anything that looks that heavy that it would justify this early optimisation. It feels like going back to normal instantiation would not impact much.

AlexMisiulia: Just good practice. If the object can be reused between the screens why should I create it each time? Semantically it is singleton because it stores the state which is common to the whole app.

cquemin: Ok if semantically it is I am fine. I thought it was just premature optimisation. Sometimes it is easier and cleaner to have new instances.

PR #48, opened 2019-05-16
Feature/spk 91. Create OrganizationInfo screen

AlexMisiulia:


app/src/main/java/com/yoti/mobile/android/social/keys/utils/FragmentUtils.kt
```diff
@@ -5,32 +5,29 @@ import androidx.fragment.app.Fragment
...
-    fun finishFragmentWithError(fragment: Fragment, errorResId: Int = R.string.invalid_input_params_error) {
-        Toast.makeText(fragment.activity, errorResId, Toast.LENGTH_SHORT).show()
-        val backStackEntryCount = fragment.activity?.supportFragmentManager?.backStackEntryCount ?: 0
-        // for situations when fragment wasn't added to backstack
-        if (backStackEntryCount == 0) {
-            fragment.activity?.finish()
-        } else {
-            fragment.activity?.supportFragmentManager?.popBackStack()
-        }
+fun finishFragmentWithError(fragment: Fragment, errorResId: Int = R.string.invalid_input_params_error) {
```
cquemin: This should be annotated with a `@StringRes`

app/src/main/java/com/yoti/mobile/android/social/keys/utils/FragmentUtils.kt
```diff
@@ -5,32 +5,29 @@ import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import com.yoti.mobile.android.social.keys.R
 
-object FragmentUtils {
```
cquemin: This file should be removed since it exists in the `common` module.

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    // hide previous fragment to prevent overdraw
+    val currentVisibleFragment = supportFragmentManager.findFragmentById(containerId)
+    if (currentVisibleFragment != null) {
+        transaction.hide(currentVisibleFragment)
+    }
+    transaction.commit()
+}
+
+fun finishFragmentWithError(fragment: Fragment, errorResId: Int = R.string.invalid_input_params_error) {
```
cquemin: Like mentioned on this other one we should use a `@StringRes` annotation.

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+    Toast.makeText(fragment.activity, errorResId, Toast.LENGTH_SHORT).show()
```
cquemin: We should not have a Toast not guarded by if(BuildConfig.DEBUG) because we will most likely forget about this code.

common/src/main/res/values/styles.xml
```diff
@@ -6,4 +6,41 @@
         <item name="android:fillViewport">true</item>
         <item name="android:fitsSystemWindows">true</item>
     </style>
+
+    <style name="ToolbarTitle" parent="android:TextAppearance">
```
cquemin: I would prefix all the style with `Keys.` or something similar just to make it more specific to the project and differentiate easily from when they are used in the XMLs.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
```diff
@@ -0,0 +1,58 @@
...
+        get() = _organizationDetailsLiveData
+
+    private var organizationId: Int = INVALID_ID
+
+    fun init(organizationId: Int) {
+        if(organizationId != INVALID_ID) {
+            this.organizationId = organizationId
+            loadOrganizationInfo()
+        } else {
+            _finishScreenLiveData.value = Event(R.string.invalid_input_params_error)
```
cquemin: The view model should avoid using Android dependencies. It would make more sense to expose a data class or a class with several field, or a state and the view itself should be responsible for mapping data classes/state to the right resources. Can you check that you apply this logic in all the view models?

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
```diff
@@ -0,0 +1,58 @@
...
+            organization.id,
+            organization.name,
+            organization.image,
+            organization.mainMenuButtonTextRes,
+            organization.mainMenuButtonColorRes
+        )
+    }
+}
+
+data class OrganizationDetails(
```
cquemin: This mapping with android resources should be done only in the view not in the ViewModel. The vie model should avoid having android references as much as possible.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
AlexMisiulia: But this is not Android dependency. This is a simple `Int` type (with annotation type checking as you mentioned above). It is ok to use it inside `ViewModel`. If I try to map it to `String` using `Context` that will be a problem )

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
AlexMisiulia: But this is not Android dependency. This is a simple Int type (with annotation type checking as you mentioned above). It is ok to use it inside ViewModel. If I try to map it to String using Context that will be a problem )

app/src/main/java/com/yoti/mobile/android/social/keys/utils/FragmentUtils.kt
AlexMisiulia: Done

app/src/main/java/com/yoti/mobile/android/social/keys/utils/FragmentUtils.kt
AlexMisiulia: Done

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
AlexMisiulia: Done

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
AlexMisiulia: I am not sure that I should hide this `Toast` even when debug is false. It will be really confusing to a user when the screen finished without any message. I can make error descriptions more user-friendly like "Oops, something goes wrong but we are fixing it".

common/src/main/res/values/styles.xml
AlexMisiulia: Do you want to create this prefixes in `demo`,` writer/reader` app? Because I don't think that prefixes in common module are a good idea.

Like `Demo.MainMenuTitle`, `Writer.MainMenuTitle`? But demo styles are only accessible to demo app so it is impossible to use them in any other place. Maybe this verbosity is redundant in this case?

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
cquemin: In this case double check with James and Simon, but this should have been specified. The toast are probably not the best to notify the user of an unrecoverable error. Toast would be used to say, record updated, or message sent. But if there is an error and we cannot fix, neither can the user it should probably be a dialog or a fullscreen.

common/src/main/res/values/styles.xml
cquemin: I was thinking the common style should have a prefix and maybe not needed in the demo app themselves for now

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
cquemin: It is an android dependencies as it is generated by the android plugin. Think of it this way. If you were to reuse this for another platform, you would not want to have any use of R.xxx here.
Also in term of responsibilities: the view is responsible for view stuff. Mapping a state to resource is View responsibilities not ViewModel. And it is the same in MVP architecture.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
cquemin: I really strongly disagree on this one. I think the VM should expose plain old kotlin object (data classes) or even complex classes. If you follow the principle of clean architecture then The view know about the VM but not the other way around. Localised string is clearly view responsibility, not business logic. You should expose a model with or without state, the view listens to it and does the mapping.

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
AlexMisiulia: Waiting for response from them

common/src/main/res/values/styles.xml
AlexMisiulia: Like having prefix "Common" on all styles that are located in the common module?

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
AlexMisiulia: After we talk about a cross-platform development it makes sense now. This is one of the recommended approaches of displaying strings in Android [link](https://medium.com/androiddevelopers/locale-changes-and-the-androidviewmodel-antipattern-84eb677660d9).

But I agree with you and don't fully like this approach. What is the alternative for this in your opinion? I have this solution: 

- Create another abstraction (`ResourceWrapper`) and map inside this class plain kotlin constants to Android resources on view layer (Ios on another platform). But this requires more code and worth only if we have plans to reuse this business logic between two of these platforms.

AlexMisiulia: @cquemin if you have code block with an approach you like it will be very useful.

common/src/main/res/values/styles.xml
cquemin: for example. You don't like it?
It just feels like it is not really unified this way.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
cquemin: This ResouceMapper is a good idea as long as it lives in the View only. It is not only for a reusability perspective it is also from a clean Architecture perspective. your VM should only expose model classes and states that will be used by the observing view to know what mapping needs to be done. I can write an example if you really need it.

common/src/main/res/values/styles.xml
AlexMisiulia: I think that when I refactor this style I will understand that it is from the common module. I haven't watched approaches like that maybe that's why I don't like it visually). It seems not very human readable like `CommonToolbarTitle`.

Can you provide a case where this prefix is useful? During the code review maybe

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
AlexMisiulia: Smthing like that?

`class ResourceMapper {
    
    enum class ResourceData {
        // here we will store all data objects for resources, ViewModel will expose them to View
        MICRO_FINANCE_NAME
    }

    private val resourceMap = hashMapOf(
        Pair(ResourceData.MICRO_FINANCE_NAME, R.string.micro_finance_name)
    )

    // View will trigger this method to unwrap ResourceData
    fun getResource(resourceData: ResourceData) = resourceMap[resourceData]
}`

Still, don't understand the usefulness of this another abstraction. Disadvantages: I need to add a constant to this enum every time I add another resource to project which should be exposed to ViewModel. It worth it if I can reuse ViewModels between different platforms in my opinion.  

How does it help from reusability perspective? 
Agreed about Clean Architecture partially because first of all, it is Int type (not interested who generates it and that it depends on the platform).

Don't take my comment personally. I just don't find this abstraction useful.

common/src/main/java/com/yoti/mobile/android/social/common/utils/FragmentUtils.kt
AlexMisiulia: Done

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
cquemin: It is not an abstraction, I am saying that this data should not live in the ViewModel. Then you are free to implement it the way you want as long as you can only find android resources (that is what they are regardless of their type) in the View and nowhere else.

AlexMisiulia: I've created a separate task for this https://lampkicking.atlassian.net/browse/SPK-138

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/OrganizationDetailsViewModel.kt
AlexMisiulia: I've created a separate task for this https://lampkicking.atlassian.net/browse/SPK-138

common/src/main/res/values/styles.xml
cquemin: I was thinking about the doted notation : `Theme.Yoti.Splash`

AlexMisiulia: Dot notation makes things better. Ok I will do it but in a separate issue (https://lampkicking.atlassian.net/browse/SPK-157).

PR #51, opened 2019-05-17
Feature/spk 94. Create/Reuse PersonalForm screen

AlexMisiulia:


demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/DemoPersonalFormViewModel.kt
```diff
@@ -0,0 +1,113 @@
...
+        logD("DomePersonalFormViewModel", "personalFormRepository#${personalFormRepository.hashCode()}")
+        if(organizationId != INVALID_ID) {
+            loadPersonalForm(organizationId)
+        } else {
+            _errorLiveData.value = Event(R.string.invalid_input_params_error)
+        }
+    }
+}
+
+data class DemoPersonalForm(
```
cquemin: Same comment the UI specific/android platform specific field should be living in the view

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
```diff
@@ -0,0 +1,93 @@
...
+            AlertDialog.Builder(requireActivity())
+                .setTitle(errorResId)
+                .setNegativeButton(android.R.string.no, null)
+                .setPositiveButton(android.R.string.yes) { _, _ ->
+                    // just navigate to previous screen if error occurs
+                    activity?.supportFragmentManager?.popBackStack()
+                }.show()
+        }
+
+        viewModel.actionEvent.observeEvent(this) {
```
cquemin: This is too generic. Here I have no idea what type of even I am observing. We should have clearer data. And most importantly a combination of state and model class that are listened from the view as opposed to individual data that lacks context.

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
```diff
@@ -12,6 +12,7 @@ class OrganizationRepository @Inject constructor() {
         Organization(
             1,
             R.string.micro_finance_name,
+            R.string.micro_finance_form_name,
```
cquemin: A repository should never expose UI specific data, it should expose model data that are relevant to the data it can access. Here you are breaking the layers responsibilities by having your repository exposing UI specific data.
View -> ViewModel -> Repository. The arrows represents : `knows about` and they are only in one direction. So you should never had view related element in your ViewModel or in your repository.

demo/src/main/res/layout/fragment_demo_personal_form.xml
```diff
@@ -0,0 +1,143 @@
...
+            app:layout_constraintGuide_begin="@dimen/start_margin"/>
+
+    <androidx.constraintlayout.widget.Guideline
+            android:id="@+id/bottomGuideline"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal"
+            app:layout_constraintGuide_end="@dimen/bottom_margin"/>
+
+    <View
```
cquemin: Should this be a guideLine ?

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/DemoPersonalFormViewModel.kt
AlexMisiulia: But this is not Android dependency. This is a simple Int type (with annotation type checking as you mentioned above). It is ok to use it inside ViewModel. If I try to map it to String using Context that will be a problem )

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
AlexMisiulia: Agree that this is not a repository information. Maybe I can just rename this class to `OrganizationUiStorage` or something like that. The purpose of this class is to store all the information related to the UI of each organization through the whole app.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
AlexMisiulia: Done. Agree with the combination of states but it creates another problem (that the whole screen is invalidated during changing even a small part of the state). I haven't decided yet which approach works better that's why I am following the first one. What do you use in Yoti?

demo/src/main/res/layout/fragment_demo_personal_form.xml
AlexMisiulia: This view creates the background for elements, not sure that guideline can do it

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/DemoPersonalFormViewModel.kt
cquemin: As mentioned before it is and it is View responsibilities not VM

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
cquemin: You can expose different state that would be reflected in refreshing different part of your screen for example.

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
cquemin: So if this is UI related class then I would put it in a mapper. Then the View observe the VM that will expose a different model (with no UI related info, just business/sates), and the view should call the mapper.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
```diff
@@ -72,7 +72,7 @@ class DemoPersonalFormFragment : ListenableFragment<DemoPersonalFormFragment.OnF
                 }.show()
         }
 
-        viewModel.actionEvent.observeEvent(this) {
+        viewModel.returnToMenuEvent.observeEvent(this) {
```
cquemin: You should not go through the VM just because a button has been clicked, this is more a presenter kind of logic. This is an unnecessary indirection. Call you listener straight from you actionButton click event.
Make sure you remove this type of wiring as they mainly pollute the VM without adding much value.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
AlexMisiulia: Sounds interesting, I need to dive deeper into this topic too as with Navigation component. Don't think that a current project is a good candidate for this

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
AlexMisiulia: The same discussion as with having android Resources inside ViewModel. After we solve it I can move to this comment

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
AlexMisiulia: @cquemin Can't agree on this. I like the "single source of truth" approach. Now it seems that it adds more level of complexity but in the future can be cases when I need show error instead of navigating to the main menu if *some condition*. That's why I think that this approach is good in a long-term perspective.

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/DemoPersonalFormViewModel.kt
AlexMisiulia: I've created a separate task for this https://lampkicking.atlassian.net/browse/SPK-138

demo/src/main/java/com/yoti/mobile/android/social/demo/repository/OrganizationRepository.kt
AlexMisiulia: I've created a separate task for this https://lampkicking.atlassian.net/browse/SPK-138

demo/src/main/java/com/yoti/mobile/android/social/demo/features/keydemo/PersonalFormFragment.kt
cquemin: Ok I am fine with it.

cquemin:
Since there is another PR this one is good to go

PR #53, opened 2019-05-20
refactoring/spk-135. Make field "Full name" mandatory during "Create form flow"

AlexMisiulia:


app/src/test/java/com/yoti/mobile/android/social/keys/features/writekey/TakePhotoViewModelTest.kt
```diff
@@ -112,7 +112,7 @@ class TakePhotoViewModelTest : BaseViewModelTest() {
         viewModel.onTakePhotoClicked()
 
         val bitmap = mock(Bitmap::class.java)
-        whenever(imageUtils.toBitmap(file.absolutePath)).thenReturn(bitmap)
+        whenever(imageUtils.toBitmapNotNull(file.absolutePath)).thenReturn(bitmap)
```
cquemin: What does this toBitmapNotNull do? how does it cope for the case where there is no bitmap?

AlexMisiulia: It throws an error. This is similar to `requireActivity()` or `requireContext()`. You should call this method if you're sure that bitmap exists for this path

PR #57, opened 2019-05-22
refactoring/spk-138. Replace Android resources in Presentation, Domain layer with ResourceData

AlexMisiulia:


common/src/main/java/com/yoti/mobile/android/social/common/features/shared/ui/resource/ResourceMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+object AndroidResourceMapper {
+
+    fun getString(activity: Activity, resourceData: StringResource) : String {
+        val resourceId = getStringRes(resourceData)
+        return activity.getString(resourceId)
+    }
+
+    @StringRes fun getStringRes(resourceData: StringResource): Int {
+        return CommonResourcesStorage.stringResourcesMap[resourceData]
+            ?: throw IllegalArgumentException("Can't find resource for $resourceData. Add it to AndroidResourceMapper")
```
AlexMisiulia: Move this to the common method

AlexMisiulia:
@cquemin You responded me in the RocketChat that you have a better implementation in your head. 

> quemincy
> 4:08 PM
> Hey Alex, i am sorry if I wasn't clear enough
> 
> May 23, 2019 10:25 AM
> but indeed this implementation is way to complex
> adds uncessary objects and singleton
> let me try to show you with 2 classes (view and ViewModel ) what I meant

I don't like the current solution because it adds complexity.

cquemin:
Indeed you are right this adds way too much complexity.
here is what I had in mind. In your ViewModel instead of doing
`_finishScreenLiveData.value = Event(R.string.error_message_ivalid_input_parameters)` I would have expected
`_finishScreenLiveData.value = Event(Status.INVALID_INPUT_PARAMETERS)` and then in your view you subscribe to this live data and and you set your text view to the correct android resource:
```
viewModel.finishScreenLiveData.observeEvent(this) {
            when (it)
               INVALID_INPUT_PARAMETERS ->  finishFragmentWithError(this, R.string.error_message_ivalid_input_parameters)
               else -> ...
        }	        
}
```

AlexMisiulia:
@cquemin will recheck it once again. Make sense, but I am not sure about the re-usages of these statuses between screens/modules

AlexMisiulia:
I create a separate issue for this (https://lampkicking.atlassian.net/browse/SPK-158) to complete it in updated code

PR #67, opened 2019-05-30
bugfix/spk-150. "Unfortunately, Social Keys has stopped" error on saving a new field

AlexMisiulia:


cquemin:
Is that the right fix for a crash ? Can you make sure you explained in here or in the ticket and more importantly in the commit how this is fixing a crash? 
We need to make sure we always explain why, the reason will be lost in few weeks otherwise.

AlexMisiulia:
@cquemin I wrote at the end of the commit message: `because Android system can't find next focusable View`.

Anyway I added the comment to the Jira issue: https://lampkicking.atlassian.net/browse/SPK-150

cquemin:
Ok Cool but this would have been helpful in the PR description as well.

AlexMisiulia:
Cyrille Quemin asks me to add an explanation for this issue. Here is the [link](https://stackoverflow.com/questions/13614101/fatal-crash-focus-search-returned-a-view-that-wasnt-able-to-take-focus/47991577#47991577) to stackoverflow question. Brief:

This error occurs when the ImeOptions is set to EditorInfo.IME_ACTION_NEXT or EditorInfo.IME_ACTION_PREVIOUS. While the parent of that view is not focusable, or the next found focus object is not focusable.

It calls to find the next focus item that should be quickly jumped to but this does not exist or isn't focusable, it throws this error. It will happen if the next is hidden and thus not focusable or the parent cannot actually deal with the findFocus() call and returns null.

The solution thusly is rather easy. Don't set it to IME_ACTION as next in this case. If you can't quickly jump from text field to text field then either make that doable or switch the action to being DONE.

PR #72, opened 2019-06-03
feature/spk-152. CR - add error prompt for expired forms (key expiry date in the past)

AlexMisiulia:


app/src/main/java/com/yoti/mobile/android/social/keys/features/writekey/WriteKeyActivity.kt
```diff
@@ -43,8 +45,12 @@ class WriteKeyActivity :
         }
     }
 
-    override fun onFormSelected(form: UiFormItem) {
-        navigateTo(TakePhotoFragment.newInstance(form.projectName, form.name))
+    override fun onFormSelected(form: Form) {
```
cquemin: This almost feels like this logic should be handled in the viewModel. The view should not need to take decision if can be avoided. The VM should expose exactly a model that the view needs to be able to be rendered.

AlexMisiulia: This is the communication between the `Fragment` -> `Activity`. Moreover: this Fragment is reused between the different activities. Now they need the same data to start the next screen, but in the future, it can be changed.

cquemin: I was talking about this bloc of code : 
```
if (isKeyExpired(form.keyExpireDateMillis, Date())) {
            showInfoDialog(R.string.write_key_expired_key_error, this)
        } else {
            navigateTo(TakePhotoFragment.newInstance(form.projectName, form.name))
        }
```

the `if` bloc should probably be handled inside the ViewModel

AlexMisiulia: Done

PR #81, opened 2019-06-21
SPK-163. Numeric keypad doesn't appear on PIN re-entry screen after attempting to read key with incorrect PIN

AlexMisiulia:


common/src/main/java/com/yoti/mobile/android/social/common/features/shared/activity/FragmentHostActivity.kt
```diff
@@ -0,0 +1,22 @@
+package com.yoti.mobile.android.social.common.features.shared.activity
+
+import android.os.Bundle
+import androidx.annotation.IdRes
+import androidx.fragment.app.Fragment
+import com.yoti.mobile.android.social.common.features.shared.fragment.FragmentNavigationListener
+import com.yoti.mobile.android.social.common.features.shared.fragment.FragmentNavigationDelegate
+
+abstract class FragmentHostActivity : BaseActivity(), FragmentNavigationListener {
```
cquemin: If you took the time to refactor that, I would have thought you would have implemented the navigation graph ?

AlexMisiulia: No, this is just to have a callback when Fragment became visible/invisible to the user. It doesn't take time because it is copy paste from my pet project

cquemin: ok. It is fine for now then. But be sure you are not adding more and more functionality and abstractions to this as the Navigation framework takes care of most of it.

PR #87, opened 2019-07-25
Bugfix/spk 169. Optimize logic of writing only lastVisitDate instead of the whole object during the read flow.

AlexMisiulia:


common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
```diff
@@ -19,20 +19,25 @@ private const val TAG = "NfcDataConverter"
...
-    fun <T> toKeyPayload(entity: T): IKeyPayload<GenericAttributeType>? {
+    fun toKeyPayload(nfcPersonalFormData: NfcPersonalFormData): IKeyPayload<GenericAttributeType>? {
         try {
             val genericPayloadBuilder = KeyPayloadBuilderFactory().createGenericPayloadBuilder()
             val versionFile = getVersionDescriptorFile()
             genericPayloadBuilder.addAttributeDataToKeyPayload(versionFile, ENTITY_STRUCTURE_VERSION.toByteArray())
 
             val entityFile = getEntityDescriptorFile()
-            val entityBytes = serialize(entity)
+            val entityBytes = serialize(nfcPersonalFormData)
```
cquemin: We really need to look at a different way to serialise data !!!!
Can you please create a ticket for this.

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
```diff
@@ -42,56 +51,73 @@ class AndroidNfcInteractor @Inject constructor(
...
 
-            val nfcDataKeyPayload = nfcDataConverter.toKeyPayload(data)
             val writeKeyRequest = WriteKeyInteractor.WriteKeyRequest(nfcTag, nfcDataKeyPayload as IKeyPayload<Any>)
                 .apply { checkGenuinity = checkGenuine }
             writeKeyInteractor.run(writeKeyRequest)
         }
     }
 
-    override suspend fun loadConfig() = withContext(ioDispatcher) {
+    override  suspend fun loadDefaultConfig() = loadConfig(nfcDataConverter.getAllFilesConfig())
```
cquemin: I can see you are calling the loadConfig before read and write. Which is fine (although you could do it only once) but more importantly this is a suspend function so it will be executed potentially in parallel whilst the rest of the read/write operation will carry on. If you want to call it each time before actually performing the read/write then look at the `SampleKeyPresenter` that shows how this could be done.

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
```diff
@@ -42,56 +51,73 @@ class AndroidNfcInteractor @Inject constructor(
...
-        data: T,
+        nfcDataKeyPayload: IKeyPayload<GenericAttributeType>?,
+        configFilesList: List<TagDescriptor.File>,
         resetPrevData: Boolean,
         checkGenuine: Boolean
     ) {
         withContext(ioDispatcher) {
 
-            loadConfig()
+            loadConfig(configFilesList)
```
cquemin: I can a `loadConfig` and a `loadDefaultConfig`. 
How does this work if a key comes from another tablet ? Will it be able to have the correct config?
What is the difference between `loadConfig` and `loadDefaultConfig`?

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
```diff
@@ -41,32 +50,52 @@ class NfcDataConverter @Inject constructor() {
...
             .shouldBeEncryptedIndividually(false)
             .genericAttributeType(GenericAttributeType.STRING)
             .build()
     }
 
-    fun <T> fromKeyResponse(response: ReadKeyInteractor.ReadKeyResponse?): T? {
-        return try {
-            val keyData = response?.keyData as GenericKeyData?
-            val entityInfo = keyData?.getFile(ENTITY_ATTRIBUTE_NAME)?.getRawValue()
+    private fun getEntityUpdatableInfoDescriptorFile(): TagDescriptor.File {
```
AlexMisiulia: Too long naming, can use just `getUpdatableInfoFile`

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
```diff
@@ -19,25 +22,33 @@ private const val TAG = "NfcInteractor"
...
+    fun getNfcMsgSize(data: NfcData): Int?
 }
 
 @Singleton
 class AndroidNfcInteractor @Inject constructor(
     private val nfcDataConverter: NfcDataConverter,
-    private val readKeyInteractor: ReadKeyInteractor,
-    private val writeKeyInteractor: WriteKeyInteractor,
+    private val readKeyInteractor: ReadKeyPayloadInteractor,
+    private val writeKeyInteractor: WriteKeyPayloadInteractor,
```
AlexMisiulia: Rename to `writeKeyPayloadInteractor`

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
```diff
@@ -74,23 +84,42 @@ class AndroidNfcInteractor @Inject constructor(
             }
 
             val nfcDataKeyPayload = nfcDataConverter.toKeyPayload(data)
-            val writeKeyRequest = WriteKeyInteractor.WriteKeyRequest(nfcTag, nfcDataKeyPayload as IKeyPayload, checkGenuine, pin)
+            val writeKeyRequest = WriteKeyPayloadInteractor.WriteKeyPayloadRequest(nfcTag, nfcDataKeyPayload as IKeyPayload, checkGenuine, pin)
```
AlexMisiulia: @cquemin names are too long, I understand that it is easier to get Context from this but maybe `WriteKeyPayloadRequest` without outer class is describable? Just for future library releases)

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/shared/BaseNfcViewModel.kt
```diff
@@ -51,6 +52,12 @@ abstract class BaseNfcViewModel(viewModelScope: CoroutineScope) : BaseViewModel(
         val nfcException = throwable as? NfcException ?: return NfcScreenError.GENERAL_ERROR
         logError(TAG, "mapToNfcScreenError, reason = ${nfcException.reason}")
 
+        // now SDK sends error with cause InvalidResponseLengthException (OUT of EPROM, Auth error)
```
AlexMisiulia: @cquemin maybe it is better to encapsulate this logic inside SDK? What do you think?

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
AlexMisiulia: Are you talking about replacing java serialization with protobuff or another serialization mechanism?

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
AlexMisiulia: > but more importantly this is a suspend function so it will be executed potentially in parallel whilst the rest of the read/write operation will carry on

I think that it will be executed sequentially one operation after another on `Worker` thread because loadConfig is blocking under the hood. I think that we don't have any problems with concurrency here.

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
AlexMisiulia: This was just my fantasy how can I write one file loading only it to config. Not actual now

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
AlexMisiulia: Done

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
AlexMisiulia: Done

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
cquemin: Yes I think we should use protobuf serialisation since the size is much better but this is not a blocker

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
cquemin: You are right if they are on the same coroutine context then they should be sequential

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
cquemin: You could a static import of this class since the static class does not add any information

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/shared/BaseNfcViewModel.kt
cquemin: I agree.

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
AlexMisiulia: I've created a ticket for that: https://lampkicking.atlassian.net/browse/SPK-189

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcDataConverter.kt
```diff
@@ -41,32 +50,52 @@ class NfcDataConverter @Inject constructor() {
         }
     }
 
-    fun getVersionDescriptorFile(): TagDescriptor.File {
+    fun fromKeyResponse(response: ReadKeyPayloadInteractor.ReadKeyPayloadResponse?): NfcData? {
```
cquemin: Static import here as well

common/src/main/java/com/yoti/mobile/android/social/common/features/nfc/data/NfcInteractor.kt
```diff
@@ -74,23 +84,42 @@ class AndroidNfcInteractor @Inject constructor(
...
+        data: NfcUpdatableInfo,
+        pin: String,
+        checkGenuine: Boolean
+    ) {
+        withContext(ioDispatcher) {
+
+            loadConfig()
+
+            val genericFile = nfcDataConverter.getUpdatableGenericFile(data)
+            val writeKeyFileRequest = WriteKeyFileInteractor.WriteKeyFileRequest(
```
cquemin: "static import"

PR #89, opened 2019-07-25
SPK-184. Can not write data if key contains corrupted data (because of unsuccessful previous write)

AlexMisiulia:


app/src/main/java/com/yoti/mobile/android/social/keys/features/writekey/WriteNfcViewModel.kt
```diff
@@ -92,6 +99,19 @@ class WriteNfcViewModel @Inject constructor(
         }
     }
 
+    private suspend fun tagContainsNfcPersonalData(nfcTag: AbstractNFCTagDefinition<Any>) : Boolean {
```
cquemin: So technically you will have 2 full read operations ? the one you do to get the data initially and the one you do here? is that correct?

AlexMisiulia: No. Before I just perfom only write operation and if error KEY_NOT_EMPTY I showed a user error msg. But I saw cases when key has corrupted data (I guess it can be during previous unsuccessful write operation), that's why it sends the same error KEY_NOT_EMPTY which is true but not good for our case.

That's why I decide to read information from the key and only if it contains not corrupted NfcPersonalFormData object I will show an error. Does it make sense?

cquemin: Yeah it makes so in most cases it will be that the key is empty so not everything is read.
When it is NOT empty then it makes sense to read everything and erase the content if required.

PR #92, opened 2019-08-02
SPK-192. Throw NOT_ENOUGH_STORAGE error at app side, fix error messages

AlexMisiulia:


AlexMisiulia:
> I will eventually move those in the SDK but good for now.

I've created a ticket for this: https://lampkicking.atlassian.net/browse/SPK-194

PR #95, opened 2019-09-11
SPK-210: rotate image from camera

cquemin:
Update for changes coming from the new key SDK 1.4.1
Changes to rotate the picture from the camera when needed.

@lampkicking/android-dev


common/src/main/java/com/yoti/mobile/android/social/common/utils/ImageUtils.kt
```diff
@@ -83,7 +87,27 @@ class ImageUtils @Inject constructor(private var ioDispatcher: CoroutineDispatch
...
+            ORIENTATION_ROTATE_90 -> 90f
+            ORIENTATION_ROTATE_180 -> 180f
+            ORIENTATION_ROTATE_270 -> 270f
+            ORIENTATION_NORMAL -> 0f
+            else -> 0f
+        }
+
+        Matrix().apply {
+            postRotate(angle)
+            return Bitmap.createBitmap(currentBitmap, 0, 0,
```
damienlo: this could throw a OOM exception maybe we can add a @throws (for java compatibility)

cquemin: This code base is entirely in Kotlin so I would avoid trying to add java compatibility since we are trying to move away from Java :)

damienlo: fait point :shipit:

marcoti:
Cause i don't have write access to this repo, i can't create my own label to add to your PR
but consider this labelled 🏷

PR #97, opened 2019-09-13
SPK-217: show keyboard when fragment become visible

cquemin:
All in the title.
@lampkicking/android-dev


common/src/main/java/com/yoti/mobile/android/social/common/features/pincode/PinCodeFragment.kt
```diff
@@ -118,6 +115,22 @@ class PinCodeFragment : ListenableFragment<PinCodeFragment.OnPinCodeEnteredListe
         }
     }
 
+
+    override fun onHiddenChanged(hidden: Boolean) {
```
damienlo: Just to be sure, according to the [documentation](https://developer.android.com/reference/android/app/Fragment.html#onHiddenChanged(boolean))

> ... Fragments start out not hidden; this will be called whenever the fragment changes state from that.

We might want to add an extra  `showSoftKeyboard` in the onViewCreated

cquemin: You are right, there is already a call in the method `initView` which is called inside `onViewCreated`

app/src/main/java/com/yoti/mobile/android/social/keys/features/writekey/WriteNfcViewModel.kt
```diff
@@ -78,24 +76,31 @@ class WriteNfcViewModel @Inject constructor(
...
-        // need to actually check if key contains not corrupted NfcData object
-        // because if key was removed during write, it can store some corrupted data that we should ignore and rewrite
-        if(tagContainsNfcPersonalData(nfcTag)) {
-            setNfcError(NfcScreenError.KEY_NOT_EMPTY)
-            return
-        }
+        try {
+            // need to actually check if key contains not corrupted NfcData object
+            // because if key was removed during write, it can store some corrupted data that we should ignore and rewrite
+            if (tagContainsNfcPersonalData(nfcTag)) {
```
RobertLK: This method can't throw because it already has its own `catch (Throwable)` inside. Need to carefully define what we expect to throw and why because having multiple layers of catching everything and trying to handle it just makes the errors really subtle and hard to find (as we've seen with all the crypto crashes in the app).

And see https://github.com/lampkicking/yoti-tag-android/pull/80 for why catching Throwable is a bad idea.

app/src/main/java/com/yoti/mobile/android/social/keys/features/writekey/WriteNfcViewModel.kt
```diff
@@ -114,7 +114,7 @@ class WriteNfcViewModel @Inject constructor(
             nfcInteractor.read(nfcTag, pinCode)
         } catch (throwable: Throwable) {
             logError(TAG, "tagContainsNfcPersonalData, read error", throwable)
-            null
+            throw throwable
```
RobertLK: Should remove this try/ catch entirely because all it's doing is logging which will just lead to multiple log spams for each issue.

PR #98, opened 2019-10-02
SPK-239: Impossible to write on an empty key (!)

cquemin:
The issue was, before writing to key we attempt to read it to see if it contains data.
When a key is empty, the Read SDK interactor returns `KEY_EMPTY` (rightly so). This exception was simply thrown like any other during the writing flow.

The fix consist of catching this particular exception with this exact reason which in turn should make the method `tagContainsNfcPersonalData` return false.

@lampkicking/android-dev


RobertLK:
Two concerns here:
* Why is 'read` nullable if it throws on empty key? Feels like this issue would not have happened if there was documentation describing these fail states.
* Why have no tests been updated?

cquemin:
These are valid point @RobertLK. 
1) I dont know why it is nullable you are right I will update the code
2) I thought they were no unit tests for this class, but I didnt look properly and there is. Will fix.

PR #99, opened 2019-10-16
SPK-222:Part 1: Update Dagger conf to reflect best practices in Yoti app

cquemin:
SPK-222 requires to create a couple of new interactors and the way Dagger was setup prevented me from adding those dependencies easily. Since I needed to fix the current Dagger implementation first I figured I should align it to how we now do things in Yoti app.

The current paradigm is followed:

- ApplicationComponent depends on:
- ApplicationModule : with Singletons
- One module per activity/feature (read tag, write tag, create form, etc.)

The feature module are defined this way:

- A module that is technically a submodule using those annotation: `@FeatureScope ` 
-  and `@ContributesAndroidInjector`
- The method annotated with `@ContributesAndroidInjector` declares dependency on the fragments specific modules
- Each Fragment Module then declare method tp provides the fragment annotated with `@ContributesAndroidInjector` and has the required dependencies in its modules annotation

@lampkicking/android-dev


app/src/main/java/com/yoti/mobile/android/social/keys/db/FormCustomFieldListConverter.kt
```diff
@@ -4,10 +4,10 @@ import androidx.room.TypeConverter
 import com.google.gson.Gson
 import com.google.gson.reflect.TypeToken
 import com.yoti.mobile.android.social.common.entity.FormCustomField
-import com.yoti.mobile.android.social.keys.di.Injector
 
+// This class needs an empty constructor to be used as `TypeConverters`
 class FormCustomFieldListConverter {
-    private val gson: Gson = Injector.appComponent.getGson()
+     val gson = Gson()
```
RobertLK: Can't we make this injected?

cquemin: No we can't really easily because this is instantiated by Gson engine. I would try to have a field injection but I really don't think it worth it. I was chatting with @damienlo about this and in the end we agreed that the effort is not worth the benefit.

cquemin: No we can't really easily because this class is instantiated by the Gson engine and therefore needs an empty default constructor. I have discussed this with Damien and perhaps I could use a field injection but we agree that the effort is not worth it.

app/src/main/java/com/yoti/mobile/android/social/keys/di/ApplicationModule.kt
```diff
@@ -0,0 +1,35 @@
...
+@Module(includes = [AndroidInjectionModule::class])
+class ApplicationModule {
+
+    @Singleton
+    @Provides
+    fun providesDatabase(application: SocialKeysApplication): AppDatabase {
+        return Room.databaseBuilder(application, AppDatabase::class.java, DB_NAME).build()
+    }
+
+    @Singleton
```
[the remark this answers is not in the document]
cquemin: That is a good point but I was thinking more about semantically that is what we want. If we change the underlying implementation then or code still behave as expected. I would have thought that we should go for the behaviour that we want as opposed to dig deep in the implementation no?

app/src/main/java/com/yoti/mobile/android/social/keys/features/shared/features/formlist/di/FormListFragmentModule.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.yoti.mobile.android.social.common.di.ViewModelFactoryModule
+import com.yoti.mobile.android.social.common.di.ViewModelKey
+import com.yoti.mobile.android.social.keys.features.shared.features.formlist.FormListFragment
+import com.yoti.mobile.android.social.keys.features.shared.features.formlist.FormListViewModel
+import dagger.Binds
+import dagger.Module
+import dagger.android.ContributesAndroidInjector
+import dagger.multibindings.IntoMap
+
+@Module(includes = [CoroutinesModule::class, ViewModelFactoryModule::class])
```
RobertLK: As discussed, this seems a like we're still hardwiring specific implementations together, which seems against the Dagger philosophy. Think these core things should be moved up to the AppComponent to make them easier to change and Dagger can pass them down to the modules.

cquemin: I totally agree with you.

app/src/main/java/com/yoti/mobile/android/social/keys/di/ApplicationModule.kt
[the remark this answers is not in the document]
DGuerrero: Yep, I agree @cquemin , I think we still should define the scope of the dependency no matter of the implementation.

app/src/main/java/com/yoti/mobile/android/social/keys/features/shared/features/formlist/di/FormListFragmentModule.kt
DGuerrero: Agreed

app/src/main/java/com/yoti/mobile/android/social/keys/di/ApplicationModule.kt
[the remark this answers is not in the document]
RobertLK: Yeah, doesn't really make a difference as long as `formDao()` always returns the same thing. If this can change we may get stale cached versions. I assume this is the case, but not familiar enough with Room so worth double checking

app/src/main/java/com/yoti/mobile/android/social/keys/di/ApplicationModule.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.yoti.mobile.android.social.keys.db.FormDao
+import dagger.Module
+import dagger.Provides
+import dagger.android.AndroidInjectionModule
+import javax.inject.Singleton
+
+@Module(includes = [
+    AndroidInjectionModule::class,
+    CoroutinesModule::class,
+    ViewModelFactoryModule::class
```
RobertLK: Shouldn't all of these actually be at the `ApplicationComponent` level?

DGuerrero: Yep, this should go in the Component.

cquemin:
Thanks a lot @DGuerrero for your comments. I agree with you there are some inconsistencies in this repo. I will create tickets to address them so that the code is totally aligned is the CA approach. But In term of Dagger implementation I assume I can merge this PR.

